Issue #85, opened 2016-04-26
Ignoring changes when calling changeState!


malx122:
We have a "command" client custom element that takes a couple of different  commands from the server and acts on the client.
One of the command is `morph_url` that tells the client to call `puppet.network.changeState` to a provided url.

However, the custom element also resets the command parameter to `''` in order to be able to receive the exactly same command again. But that change is never sent to server causing the servers and clients data model to differ!

Pseudo code in custom element, `command` is bounded to the"`Model.Command$`

```
commandChanged: function () {
   parsecommand();
   this.set('command', '');
   window.tu.puppetDirtyCheck(); (Utilityfunction to force the puppet  check that otherwise only reacts on click)
   ....
  ....
   puppet.network.changeState(parameterfromcommand);
          
}

```

miyconst:
Do you have `notify: true` in the `command` property declaration?

malx122:
Yes! As I said, the client model is updated. Its just not sent to server.

miyconst:
Am I right, that the change is not being sent to the server even with the next patches from the client side?

malx122:
Yepp, thats right. The client and server model differs even after a couple of patches.

miyconst:
That's definitely an error then.

@malx122 if you could create a small example which demonstration of the issue, it would be very useful.
@warpech take a look please.

malx122:
I think the first post describing the problem good enough. If you have trouble to repeat it locally, call me and I will show it.

warpech:
Does it work when you click on the page background or trigger window `mouseup` event as described here? https://github.com/PuppetJs/PuppetJs#generating-patches-based-on-local-changes

warpech:
Aha, I guess this is what `window.tu.puppetDirtyCheck();` does.

warpech:
Can you reproduce it in a small app, like https://github.com/warpech/SimplePuppet or https://github.com/StarcounterSamples/KitchenSink?

malx122:
Not sure, Ivar found out that it works if he wraps the set command in an async:
(which of course should not be needed)
```
this.async( function() {
     this.set('command', '');
     window.tu.puppetDirtyCheck();
 }).bind(this);
```

malx122:
The whole problem is that you "unobserve" the model while applying data patches from server. That is not ok since applying data patches may trigger javascript methods that makes other model changes themself. 

[...]

PR #88, opened 2016-06-20
Return results from jsonpatch.apply; call onRemoteChange outside observe. #85

joozek78:


joozek78:
depends on starcounter-jack/JSON-patch#110

src/puppet.js
```diff
@@ -471,6 +471,18 @@
 
   Puppet.prototype = Object.create(EventDispatcher.prototype); //inherit EventTarget API from EventDispatcher
 
+  Puppet.prototype.dispatchErrorEvent = function(error) {
```
tomalec: Do we need it in public API?

src/puppet.js
```diff
@@ -560,44 +572,52 @@
...
+      } else {
+        throw error;
       }
     }
-    else {
-      this.jsonpatch.apply(tree, sequence);
+
+    var that = this;
+    sequence.forEach(function (patch) {
+      if (patch.path === "") {
```
tomalec: @warpech do we need this check in non-debug mode?

src/puppet.js
joozek78: No, it's just my inexperience with JS showed its ugly head :) I'll move it

src/puppet.js
```diff
@@ -622,40 +642,18 @@
...
 
     if (this.onPatchReceived) {
         this.onPatchReceived(data, url, method);
     }
 
-    if (!this.observer) {
-      return; //ignore remote change if we are not watching anymore
-    }
+    // apply only if we're still watching
+    if (this.observer) {
```
tomalec: Personally I prefer "return fast" in simple cases, as then I do not have to read entire method, and search for applicable `else`

test/spec/OnLocalRemoteChangeSpec.js
```diff
@@ -60,11 +58,40 @@ describe("OnLocal/RemoteChange", function() {
...
                 "op": "replace",
                 "path": "/hello",
                 "value": "onRemoteChange callback"
             }]);
+            expect(mostRecentCall.args[1]).toEqual([{removed: 'world'}]);
+            done();
+        }, 100); // for FF
+    });
+
+    it("should fire patch-applied event for applied patches", function(done) {
```
tomalec: Could you also add a test  for `should NOT fire patch-applied event for received, and queued patches`?

src/puppet.js
joozek78: I just find it clearer and more intuitive, but I don't have to stick to it.

tomalec: For me it's less intuitive, as to preserve democracy, you can ask Warpech ;)

Issue #71, opened 2015-08-24
Session reconnection in case of lost response


warpech:
As discussed in https://github.com/PuppetJs/PuppetJs/issues/37#issuecomment-112757820 downwards, there may be a time where server response (e.g. `_ver#s: 2`) is irreversably lost due to network error. 

This is a case when the server created a message `_ver#s: 2`, but it never reached the client. In result, the following server responses `_ver#s: 3`, `_ver#s: 4` will queue up on the client forever, while the client still is waiting for `_ver#s: 2`.

The solution should enable a nice Gmail-style reconnection message, as explained here: http://ux.stackexchange.com/questions/83371/web-app-is-offline-show-reconnecting-in-countdown

![](http://i.stack.imgur.com/pWgVX.png)

@tomalec could you wrap up the proposed solution?

This is a long term stability improvement, no need to do it immediately.

warpech:
Related: https://github.com/Starcounter/Starcounter/issues/3084

tomalec:
Proposition for 

## Session re-establish flow

1. Local changes are made:
 1. PuppetJS sends messages
 2. `puppet.queue` adds versions, etc
 3. `puppet.queue` typeof `-OT-agent` keeps a list of  `.pending/.pendingVersionedPatches` with all Versioned-JSON-Patches
as well as (at least for internal use) `.pendingMessages/.pendingPatches`

2. Somebody decides to re-establish connection, due to WS timeout, connection lost, whatever
  ### A - local decision
 4. Puppet re-establishes connection, with some flags, that this is "re-established", not new connection, and with most recent state (versions) of queue
 5. Puppet sends once again all `.queue.pending` changes

 ### B remote decision 
 4. Remote (for example Starcounter Puppeteer) reestablished connection with its local, remote, ack versions
 5. Remote sends all its pending patches
 6. Puppet calls `.queue.receive( allRemotesPendingPatches, silent = true)`
so queue will silently disregards already applied patches

warpech:
Good idea! What we are still missing is to describe the protocol for "re-establishing" the connection. The primary work needs to be done in the [spec](https://github.com/PuppetJs/PuppetJs/wiki/Server-communication:-advanced-synchronization), then agreed with @chrhol and finally implemented.

There is no chance of finishing it this year, so let's get back to it in January.

dan31:
This is the main issue to track session reconnection issue. CC @malx122.
To finish the work, @warpech assgined @joozek78.

[...]

PR #90, opened 2016-07-07
Reconnection in case of lost connection. #71

joozek78:


joozek78:
I still have to connect the count-down from [here](https://github.com/PuppetJs/PuppetJs/pull/90/files#diff-313238889fca7bf671d9b24abf12eb2eR115) to puppet-client somehow.
CC @tomalec @warpech 
See also PuppetJs/JSON-Patch-OT-agent#1

src/puppet.js
```diff
@@ -89,6 +89,122 @@
...
+   * Resets state of Reconnector
+   */
+  Reconnector.prototype.notifySuccessfulConnection = function() {
+    this._reset();
+  };
+
+  /**
+   * Guarantees some communication to server and monitors responses for timeouts.
+   * @param sendHeartbeatAction will be called to send a heartbeat
+   * @param onError will be called should a response fail to arrive in `timeoutMs`
```
tomalec: I don't get this description

src/puppet.js
```diff
@@ -387,51 +538,38 @@
...
-      if (puppet.debug) {
-        puppet.remoteObj = responseText; // JSON.parse(JSON.stringify(puppet.obj));
-      }
-
-      recursiveMarkObjProperties(puppet.obj);
-      puppet.observe();
-      if (onDataReady) {
-        onDataReady.call(puppet, puppet.obj);
+    this.onDataReady = options.callback;
+    this.network.baseRemoteUrl = this.network.remoteUrl;
```
tomalec: What it is for?

src/puppet.js
joozek78: will be called if (after a message has been sent) no response will arrive in `timeoutMs`. how about now?

src/puppet.js
```diff
@@ -653,6 +780,11 @@
     }
 
     this.queue.receive(this.obj, patches);
+    if(this.queue.pending && this.queue.pending.length && this.queue.pending.length > 1) {
```
tomalec: This number should not be hard-coded here.

Plus, I do not think `1` is the good number. I'm afraid it may interrupt our keystroke integration.

src/puppet.js
```diff
@@ -89,6 +89,122 @@
...
+  /**
+   * Guarantees some communication to server and monitors responses for timeouts.
+   * @param sendHeartbeatAction will be called to send a heartbeat
+   * @param onError will be called should a response fail to arrive in `timeoutMs`
+   * @param intervalMs if no request will be sent in that time, a heartbeat will be issued
+   * @param timeoutMs shuold a reponse fail to arrive in this time, `onError` will be called
+   * @constructor
+     */
+  function Heartbeat(sendHeartbeatAction, onError, intervalMs, timeoutMs) {
+    this._send = sendHeartbeatAction;
```
tomalec: Personally, I do not like so many `._pretendToBePrivate` and I would either made them public or really private

joozek78: well, I didn't want to isolate them in IIFEs, because it would seem strange in this codebase. Underscores are in our [guidelines](https://starcounter.gitbooks.io/guidelines/content/coding_style/javascript_common_practices.html)

src/puppet.js
joozek78: Agree about hard-coding. Good point about keystrokes, will test it with some textboxes

src/puppet.js
joozek78: trash

src/puppet.js
tomalec: I didn't mean IIFE but moving methods definitions into constructor, or making properties public.

Our guidelines says not to use `_` in any other case. But it does not say that we should use just naming convention instead of real protection of methods and variables.

src/puppet.js
tomalec: "will be called if no response will arrive after `timeoutMs` since a message has been sent" ?

src/puppet.js
miyconst: The thing is that there are three kind of protection levels:

- Public - available for everyone.
- Protected - available for myself and all successor classes.
- Private - available for myself only.

The `protected` mode is not supported by JavaScript, in this case an underscore (`_`) should be used. The `private` methods or variables should be physically hidden, as @tomalec said.

src/puppet.js
tomalec: What about `3` ? ;)  at least we could refer in readme to complete explanation
https://youtu.be/xOrgLj9lOwk?t=1m29s

src/puppet.js
tomalec: That's true that JS does not support `protected` in classical sense, but we can achieve similar level of protection by making them private and moving methods that uses them to constructor. We do not use and plan tu use the prototype anyways. I simply do not like abusing the conventions, as most of the members are "protected" by name.

I'd say that `onError`, `send` and `timeoutMs` could be as well complately private as public.

But it's just my personal thing.

joozek78: Sorry, I somehow missed your last comments guys.
I initially made those functions "truly private" (i.e. defined them in constructor along with public ones), but then I noticed how much this code stands out from the rest of puppet codebase (it's not used anywhere else). I can change it back, I don't care about it too much

src/puppet.js
```diff
@@ -89,7 +89,142 @@
...
+      var timeToReconnectionS = (this._timeToCurrentReconnectionMs / 1000);
+      this._dispatchEvent('reconnection-countdown', {seconds: timeToReconnectionS});
+      console.log('reconnecting in '+ timeToReconnectionS +'s');
+      this._timeToCurrentReconnectionMs -= 1000;
+      setTimeout(this._step.bind(this), 1000);
+    }
+  };
+
+  Reconnector.prototype._dispatchEvent = function (name, detail) {
+    this._eventDispatcher.dispatchEvent(new CustomEvent(name, {bubbles: true, cancelable: false, detail: detail}));
```
tomalec: does `bubbles` work even if Puppet instance does not have `.parentNode`?

joozek78: I don't know, I admit I copied it

src/puppet.js
```diff
@@ -89,7 +89,142 @@
 
   }
 
-  function PuppetNetworkChannel(puppet, remoteUrl, useWebSocket, onReceive, onSend, onError, onStateChange) {
+  /**
+   * @constructor
+     */
+  function Reconnector(eventDispatcher, reconnectAction) {
+    this._eventDispatcher = eventDispatcher;
+    this._reconnect = reconnectAction;
```
tomalec: I'd make `eventDispatcher` and `reconectAction` public members, as they are given by public API anyways, and I see no harm in changing them in runtime.

joozek78: I see no reason to change them neither. It's a lot harder to remove something from public API then to add it, so I would leave it private. If need occurs we can always add it later

tomalec:
Regarding `_` with @warpech we agreed that we should narrow the usage of it, and in general we should use it temporarily or as a last resort in few cases only. I have updated guidelines at https://starcounter.gitbooks.io/guidelines/content/coding_style/javascript_common_practices.html

So, I suggest to move methods to the constructor and remove unnecessary `_`.

----

We also agree that as in long-term solution we would like to move events from puppet.js to puppet-dom or even to `<puppet-client>`.

That's why I would suggest not to add more events here, and try to use callbacks. Puppet is supposed to work without browser; there will be nobody in between Puppet and `puppet-client` that could also subscribe to events; and `puppet-client` is EventTarget anyways so that will be the place to specify bubbling, details etc.

src/puppet.js
```diff
@@ -271,23 +271,27 @@
...
+  PuppetNetworkChannel.prototype.establish = function(bootstrap){
+    this._establish(this.remoteUrl.href, null, bootstrap);
+  };
+  PuppetNetworkChannel.prototype.reestablish = function(pending, bootstrap) {
+    this._establish(this.remoteUrl.href + "/reconnect", JSON.stringify(pending), bootstrap);
+  };
+
   // TODO: auto-configure here #38 (tomalec)
-  PuppetNetworkChannel.prototype.establish = function(bootstrap /*, onConnectionReady*/){
+  PuppetNetworkChannel.prototype._establish = function(url, body, bootstrap){
```
tomalec: Can we make it completely private?

src/puppet.js
```diff
@@ -387,51 +575,37 @@
...
-        puppet.remoteObj = responseText; // JSON.parse(JSON.stringify(puppet.obj));
-      }
-
-      recursiveMarkObjProperties(puppet.obj);
-      puppet.observe();
-      if (onDataReady) {
-        onDataReady.call(puppet, puppet.obj);
+    this.onDataReady = options.callback;
+
+    this._createQueue = function() {
```
tomalec: is there any reason in adding this function as method?

joozek78: It's jest easier to use

tomalec: How it is easier to write `this._createQueue` instead of `createQueue`, or even `` -as it could be executed inline?

Plus it makes API messy, it make another method public, and as you said, it's harder to remove things from API rather than add them.
It makes developer wonder, whether it is used ot of this scope, or by any other module.

If it is local you do not need to remember it once you have finished reading constructor.

src/puppet.js
```diff
@@ -89,7 +89,145 @@
...
+   * @constructor
+   */
+  function Reconnector(reconnect, onReconnectionCountdown, onReconnectionEnd) {
+    var intervalMs,
+        timeToCurrentReconnectionMs,
+        reconnectionPending,
+        reconnection,
+        defaultIntervalMs = 1000;
+
+    var reset = (function () {
```
tomalec: Is there any reason to keep the first parenthesis? 

Plus, I think as you are declaring them at the beginning, you can declare them as functions not function expressions.

tomalec:
Naturally we also need a note in README, brief flow description in docs and some tests.

Issue #94, opened 2016-08-31
Ping behavior


tomalec:
Bellow is a side idea of discussion on problems with heartbeats, keeping session alive, and overhead of refreshing bindings.

Personally, I find "PING" and HEAD condition confusing and sticking out from entire flow/Puppet-protocol. If I remember correctly, some time ago we talked about "touch" functionality/method in PuppetJS (https://github.com/PuppetJs/PuppetJs/issues/29) we can also add `ping` method that will send also **valid** JSON Patch, that could be easily identified as no message, will not change a thing, and will not trigger any diff calculations, but still would touch the server and session. For example:

(1.)
```
[]
```

+ It would work fine with any protocol: HTTP, WS, or even Pigeon post ;)
+ will not require any exclusions to JSON Patch parser
+ will not require any arbitrary keyword to be documented

This is the messages that works more like regular ping: send tiny message, which is not a request for any action, and just check the connection.

We can also send just one or two test operation objects, that will let both peers to acknowledge the state of the other.

(2.)
```
[
   {op: "test", path: "ServerVersion", value: 70}
]
```

(3.)
```
[
   {op: "test", path: "ServerVersion", value: 70},
   {op: "test", path: "BrowserVersion", value: 60}
]
```

They also meet the benefits mentioned above.

For heartbeats, I would use a message that really beats the rythm - counts so both peers could follow same tune:
(4.)
```
[
   {op: "test", path: "ServerVersion", value: 70},
   {op: "test", path: "BrowserVersion", value: 60},
   {op: "replace", path: "BrowserVersion", value: 61}
]
```

Such message is definitely requesting a change, so it should be no surprise that an action is taken on server, and it may produce a response.

@warpech, @chrhol, @joozek78, @miyconst, @dan31
Does it make sense to you?

Correct me if I'm wrong, but as I understood on server-side we have few actions to cover:
1. Check the connection, and bump the session,
2. Respond with diff with all the changes made since last sync,
3. Refresh everything, to reset the browser's state.
@chrhol, @joozek78 could you check that and describe the actions we need to cover?

---

Related issues:
https://github.com/PuppetJs/PuppetJs/issues/29
https://github.com/PuppetJs/PuppetJs/issues/92
https://github.com/PuppetJs/PuppetJs/issues/93

tomalec:
*In the email conversation @joozek78 wrote:*
> Regarding Tomek's idea

[...]

Issue #79, opened 2016-02-29
Deprecation proposal: Remove $parent


warpech:
PuppetJs supports special notation: a `$parent` alias. For example, the following notation:

```
puppet.obj.deep.$parent.name
```

Is equivalent to:

```
puppet.obj.name
```

As of now, it does not work with Polymer 1.2 and 1.3, because Polymer creates an explicit binding to a path `model.deep.$parent.name` and ignores changes that are set on `model.name`.

It can be fixed by normalizing Polymer bindings:

```javascript
            var templates = document.querySelectorAll("template");
            for(var i=0; i<templates.length; i++) {
                if(templates[i]._propertyEffects) {
                    for(var key in templates[i]._propertyEffects) {
                        if(templates[i]._propertyEffects.hasOwnProperty(key)) {
                            for(var j=0; j<templates[i]._propertyEffects[key].length; j++) {
                                if(templates[i]._propertyEffects[key][j].effect) {
                                    if(typeof templates[i]._propertyEffects[key][j].effect.value == "string") {
                                        if(templates[i]._propertyEffects[key][j].effect.value.indexOf(".$parent.") > -1) {
                                            templates[i]._propertyEffects[key][j].effect.value = collapseParent(templates[i]._propertyEffects[key][j].effect.value);
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
            }

    function collapseParent(str) {
        var parts = str.split(".");
        var n = parts.length - 1;
        while (n >= 0) {
            if (parts[n] == "$parent") {
                parts.splice(n-1, 2);
                n--;
            }
            n--;
        }
        return parts.join(".");
    }
```

But even better solution is to avoid this problem but stopping the support for `$parent` in PuppetJs. I propose to do this after Starcounter 2.2 is released.

tomalec:
LGTM,
I don't think we have any real use-case for `$parent`. On the other side it cost us a lot (performance, maintenance), requires more documentation, and for sure provides more magic, to otherwise pure JSON object.

warpech:
Some `$parent` is used in Administrator: https://github.com/Starcounter/Starcounter/blob/57c6d056c5a43e4ffabbea3f261e3f30219220b1/src/Starcounter.Administrator/app/partials/databaseNew.html

But I think it's Angular's, not Puppet's.

warpech:
@joozek78 please do it if you find it convenient for your current work.

[...]

PR #97, opened 2016-09-05
Feature/94 change heartbeat behavior

joozek78:
* Make heartbeats empty sequences instead of version bumps (#94)
* BTW remove support for $parent (broken commit message, sorry) #79 
* Refactor: Rename `Puppet.prototype._sendPatches` method to `sendPatches` function

### DIFF
--- a/src/puppet.js
+++ b/src/puppet.js
@@ -665,7 +665,7 @@
   Puppet.prototype.jsonpatch = global.jsonpatch;
 
   Puppet.prototype.sendHeartbeat = function () {
-    this.handleLocalChange([]); // sends empty message to server
+    this._sendPatches([]); // sends empty message to server
   };
 
   Puppet.prototype.observe = function () {
--- a/src/puppet.js
+++ b/src/puppet.js
@@ -781,7 +781,7 @@
   Puppet.prototype.handleRemoteChange = function (data, url, method) {
     this.heartbeat.notifyReceive();
     var patches = JSON.parse(data || '[]'); // fault tolerance - empty response string should be treated as empty patch array
-    if(patches.length === 0) { // heartbeat message
+    if(patches.length === 0) { // ping message
       return;
     }
 
--- a/src/puppet.js
+++ b/src/puppet.js
@@ -557,7 +557,7 @@
 
     if(options.pingIntervalS) {
       const intervalMs = options.pingIntervalS*1000;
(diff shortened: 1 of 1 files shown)


tomalec:
@joozek78 Could you briefly describe what is in the scope of this "change"?

src/puppet.js
```diff
@@ -842,7 +795,9 @@
     if(this.queue.pending && this.queue.pending.length && this.queue.pending.length > this.retransmissionThreshold) {
       // remote counterpart probably failed to receive one of earlier messages, because it has been receiving
       // (but not acknowledging messages for some time
-      this.queue.pending.forEach(this._sendPatches.bind(this));
+      this.queue.pending.forEach((function (p) {
+        sendPatches(this, p);
+      }).bind(this));
```
tomalec: If you are doing closure anyway, maybe we could get rid of `.bind` overhead?

I would either do
```js
this.queue.pending.forEach(function (p) {
    sendPatches(that, p);
});
```
or
```js
this.queue.pending.forEach(sendPatches.bind(null,this));
```

joozek78: no problem

tomalec:
What do you thing about exposing `.ping` as a method name in public API?

joozek78:
I'm not against it, but why? And don't you think it should be called `sendHeartbeat` or something like that?

tomalec:
I just think the common understanding of "ping" word is wider.

I would say that "heartbeat" is a great name for regular, rhythmic.. pinging ;)
So the 
 - "ping" is functionality to send atomic message to check the connection
 - "heartbeat" is when your ECG monitor pings regularly

joozek78:
Okay, why not. But still - why add it now? Do we have specific need for it
or do you want to provision for future needs?

tomalec:
I do not have specific need for it. Just because why not ;) as you are cleaning `$parent` you can make this little method in API shorter. Short name will align nicer with method that we will probably finally need to implement from https://github.com/PuppetJs/PuppetJs/issues/29#issuecomment-170904351

tomalec:
Other than that entire PR LGTM

src/puppet.js
```diff
@@ -828,6 +781,9 @@
   Puppet.prototype.handleRemoteChange = function (data, url, method) {
     this.heartbeat.notifyReceive();
     var patches = JSON.parse(data || '[]'); // fault tolerance - empty response string should be treated as empty patch array
+    if(patches.length === 0) { // heartbeat message
```
tomalec: should the code comment now say: "ping message"?

PR #101, opened 2016-09-13
Feature/debug fix and clean up after parent

joozek78:


joozek78:
Already fixed by tomalec

Issue #109, opened 2016-11-28
Remove cookie-based session ID setting


warpech:
Remove cookie-based session ID setting. PuppetJs should not read ID from cookies, because this leads to problems when page is loaded in many tabs.

Main issue: https://github.com/Starcounter/Starcounter/issues/3798#issuecomment-263259468

warpech:
Merged PR #111 that implements this.

PR #111, opened 2016-11-28
Breaking: Remove cookie-based session ID setting

warpech:
Implements:  #109 
Partial fix for: https://github.com/Starcounter/Starcounter/issues/3798

From now on, session ID can only be provided as `remoteUrl` option or `X-Location` header. Otherwise, the window location URL is used.


warpech:
I guess this requires increasing MAJOR component in PuppetJs version number. @tomalec do you agree?

warpech:
@tomalec can you pls proceed with the review?

src/puppet-dom.js
```diff
@@ -50,7 +35,7 @@
       this.element.removeEventListener('puppet-redirect-pushstate', this.historyHandler);
     };
 
-    options.remoteUrl = options.remoteUrl || getRemoteUrlFromCookie() || window.location.href;
+    options.remoteUrl = options.remoteUrl || window.location.href;
```
tomalec: Do we need this line, as `Puppet` constructor (`PuppetNetworkChannel`) , sets it's own default https://github.com/PuppetJs/PuppetJs/blob/e95f936484a9715694ff58f4101daa930a9c83e4/src/puppet.js#L238

If we will remove it, we will get rid of slightly redundant code, but if `remoteUrl` is not provided `puppet.remoteUrl == undefined` and `puppet.network.remoteUrl = new URL(window.location.href)`

tomalec:
I have fixed one syntax error, and add a [question](https://github.com/PuppetJs/PuppetJs/pull/111#discussion_r90643320), other than that LGTM.

src/puppet-dom.js
warpech: I agree, it looks that is is safe to remove this line completely. Great catch :) I will do it.

warpech:
Sorry for introducing syntax error. I was not able to run it without changes in Starcounter.

Here's an idea. Maybe let's use `puppetjs#remove-location-cookie` temporarily in Starcounter daily build to test this out. We can merge if this is confirmed to work.

tomalec:
:+1:

src/puppet-dom.js
warpech: Nay... Let's keep it as is, because that would be a big refactor. Out of scope in this issue.

tomalec: Fair point, but please not that we use `location` in many other places, and I would say it's quite useful to have notion of location even in node version

warpech:
Just please let me know if you agree with the last commit (https://github.com/PuppetJs/PuppetJs/pull/111/commits/0177c8ccb5e6f856c4af928cea203e883e717f8c).

tomalec:
I agree, but please check if it passes tests in all browsers.

warpech:
All tests pass in Chrome.
2 tests fail in Firefox.
7 tests fail in Edge.
1 tests fails in CLI (Shadow DOM).

This is the same result as in 2.3.0, so there is no regression in testing caused by this PR.

PR #113, opened 2016-11-29
Remove .bowerrc (use polyserve instead)

warpech:
Remove .bowerrc (use polyserve instead). Added instructions about using Polyserve to README.md

@tomalec can you pls review


tomalec:
LGTM

warpech:
Thx!

PR #202, opened 2018-02-27
`listenTo` the `document` by default,

tomalec:
as `.body` may not be there yet.

 Required for https://github.com/Starcounter/RebelsLounge/issues/248


tomalec:
I don't think we should release it withing SCF right now.

PR #231, opened 2019-04-18
Use this._attemptingScroll as attemptingScroll is undefined

tomalec:


warpech:
So, was this the source of our remaining problem https://github.com/Palindrom/Palindrom/issues/216, or is the problem still there?

PR #244, opened 2019-06-04
Improve node bundle

tomalec:
Export Palindrom as default export from node bundle,
exclude node_modules from node bundle


webpack.config.js
```diff
@@ -61,7 +63,7 @@ module.exports = [
             extensions: ['.js']
         },
         target: 'node',
-        externals: { './URLShim': 'URL' },
+        externals: [nodeExternals()],
```
warpech: How can I verify that this change is correct? The `runner-node.js` script actually checks the files in `src/`, not the Node bundle.

Before you can answer this, my real question is: What is the benefit of having a Node bundle? If I understand correctly, one good reason for using Webpack on Node is transpilation using Babel, but our tests show that this is not needed.

tomalec: > How can I verify that this change is correct? The runner-node.js script actually tests the files in src/, not the Node bundle.

Here is an issue for that https://github.com/Palindrom/Palindrom/issues/245

> Before you can answer this, my real question is: What is the benefit of having a Node bundle? 

The only reason I see is bundling all Palindrom internals (network, noqueue, reconnector etc.) into single module, to reduce friction of resolving them.

src/palindrom-network-channel.js
```diff
@@ -331,7 +331,7 @@ export default class PalindromNetworkChannel {
 
         this.onSend(data, url, method);
 
-        let isomorphicFetch = typeof global !== 'undefined' ?  global.fetch : nodeFetch;
+        let isomorphicFetch = typeof global !== 'undefined' && global.fetch || nodeFetch;
```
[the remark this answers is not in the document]
tomalec: When you run then bundle in node :)
like `npm install palindrom && node myfile.js`
where myfile.js:
```js
import Palindrom from 'palindrom';
const instance = new Palindrom({...});
```

webpack.config.js
warpech: > The only reason I see is bundling all Palindrom internals (network, noqueue, reconnector etc.) into single module, to reduce friction of resolving them.

I don't think resolving local requires is considered friction in the Node world. However, having an additional step in the build system and the complexity of debugging webpacked code is much more real friction than that.

tomalec:
Let's discuss and implement de-bundling as a separate issue. Merging

PR #250, opened 2019-06-28
Add tests for client heartbeat

tomalec:
TODO here or in a new PR:
 - tests for WS scenario,
 - tests for reconnect

PR #252, opened 2019-07-02
Move Heartbeat to PalindromNetworkChannel

tomalec:
Change in behavior:
 - do not call `unobserve`, `observe` when sending ping
 - make `Heartbeat` start after connection is established (not once JSON is parsed and queue reset)
 - notify `Heartbeat` about received change before `onPatchReceived` callback
 - removed from Palindrom API: `ping, handleConnectionError, handleFatalError`

----
requires https://github.com/Palindrom/Palindrom/pull/250 - to test affected functionality

---- 
Rationale behind this change is that ping, and keeping connection alive is related to the network channel that we are using. Also from the perspective of the core of Palindrom: observing and syncing JSON, exchanging the empty messages in intervals is not needed.


src/palindrom-network-channel.js
```diff
@@ -44,7 +45,6 @@ export default class PalindromNetworkChannel {
         onReceive && (this.onReceive = onReceive);
         onSend && (this.onSend = onSend);
         onConnectionError && (this.onConnectionError = onConnectionError);
-        onFatalError && (this.onFatalError = onFatalError);
```
warpech: Originally `onConnectionError` was used to display a message to the user saying that there are temporary connection problems but we are reconnecting. Similar to:

- image
 
On the other hand, `onFatalError` was used to display Palindrom crashes that can only be solved by reloading the page. Similar to:

- image

After this change, these two cases are undistinguished.

src/palindrom-network-channel.js
```diff
@@ -70,6 +70,18 @@ export default class PalindromNetworkChannel {
...
             }
         });
+
+        if (pingIntervalS) {
+            const intervalMs = pingIntervalS * 1000;
+            this.heartbeat = new Heartbeat(
+                () => {this.send([]);},
+                this._handleConnectionError.bind(this),
+                intervalMs,
+                intervalMs
```
warpech: I know this is copied from existing implementation, but I think it does not make sense to use the same value for timeout that is used for ping interval.

It seems better to add another configuration option for timeout or even hardcode it temporarily and make an issue to fix it in the future.

tomalec: I'd make an issue, to keep this only as refactoring.

src/palindrom-network-channel.js
tomalec: Then it seems those cases were broken before, unless I miss something.

I didn;t change the functionality, I just moved the error handling from Palindrom https://github.com/Palindrom/Palindrom/pull/252/files#diff-92b698a38d3db0381546f440a80e631bL298 
to NetworkChannel https://github.com/Palindrom/Palindrom/pull/252/files#diff-09a771a96432511bf7bc17a34547183dR119

I cannot find any other use of `palindromInstance.onFatalError` in any Palindrom custom-element, https://github.com/search?p=1&q=org%3APalindrom+onFatalError&type=Code
Also, there is no specs nor docs for `onFatalError` usage, so I considered it not public API https://palindrom.github.io/docs/search.html?q=onFatalError

src/palindrom-network-channel.js
tomalec: https://github.com/Palindrom/Palindrom/issues/254

src/palindrom-network-channel.js
warpech: I agree, let's cleanup the existence of `onFatalError` and re-add it in future in needed.

PR #253, opened 2019-07-04
Try to reduce test flakiness in Firefox by adding more time for establish

tomalec:
add note to the desc description, to emphasize waiting for establish.

### DIFF
--- a/test/specs/dom/index.js
+++ b/test/specs/dom/index.js
@@ -361,9 +361,9 @@ if (typeof window !== 'undefined') {
                 describe('palindrom-morph-url event', function() {
                     beforeEach(async () => {
                         // wait for Palindrom to call .listen (after finishing the ajax request)
-                        await sleep();
+                        await sleep(10);
                     });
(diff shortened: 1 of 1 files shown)


test/specs/dom/index.js
```diff
@@ -361,9 +361,9 @@ if (typeof window !== 'undefined') {
                 describe('palindrom-morph-url event', function() {
                     beforeEach(async () => {
                         // wait for Palindrom to call .listen (after finishing the ajax request)
-                        await sleep();
+                        await sleep(10);
```
warpech: This change actually gives **less** time for waiting, because the default sleep value is `30`. See: https://github.com/Palindrom/Palindrom/blob/a5869cd879fd7a37848f8733f5bc911db391a12e/test/utils/index.js#L1

tomalec: ha! thanks, I'll give it `100` and hope that will solve the flakiness 😕

Issue #261, opened 2019-07-16
Use `palindrom-dom` as the default entry point in Node, do not use the `browser` field


warpech:
Right now, the NPM package uses `src/palindrom` as the `main` entry point and `src/palindrom-dom.js` as the `browser` entry point.

`browser` is considered legacy and problematic. See:
- https://github.com/webpack/webpack/issues/7646 (by @webreflection)
- https://github.com/webpack/webpack/issues/4674
- https://github.com/gatsbyjs/gatsby/issues/3192

One of the problems is that `browser` is incompatible with the 3rd alternative entry point `module.` There is no 4th kind of entry point (`module-browser`).

Note: this seems to be Webpack's order of preference: `browser` > `module` > `main`.

Several other libraries have also stopped using `browser`, for example https://github.com/blockstack/blockstack.js/pull/621

tomalec:
WDYT about separating PalindromDom to a separate repo/module? It will make such problems not relevant any more.

warpech:
That would also be a solution but might backfire on us with maintenance hell.

Another solution is to keep `src/palindrom-dom.js` as the entry point, but do not use default exports. Like these guys: 

- https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/
- https://palantir.github.io/tslint/rules/no-default-export/
- https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad
- https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html

tomalec:
I vote for no "defaults".

tomalec:
Implemented in #261
Released in [@6.2.0](https://github.com/Palindrom/Palindrom/releases/tag/v6.2.0)

PR #263, opened 2019-07-16
Remove default exports (#261)

warpech:
Implements #261


warpech:
The Vue demo at https://github.com/Palindrom/palindrom.github.io/blob/ffc1db8411f0b6d33f12ac8d6ccf50129988da6b/public/lab/vue/src/App.vue#L9 is already compatible with this change, because it uses a named export.

README.md
```diff
@@ -59,14 +59,20 @@ npm install palindrom --save
...
-//CommonJS
-var Palindrom = require('palindrom');
+// CommonJS
+const Palindrom = require('palindrom').Palindrom;
+const PalindromDOM = require('palindrom').PalindromDOM;
 
 // ES6/TS
-import Palindrom from 'palindrom'
+import { Palindrom } from 'palindrom';
+import { PalindromDOM } from 'palindrom';
```
tomalec: Why not `import { Palindrom, PalindromDOM} from 'palindrom';`?

README.md
```diff
@@ -59,14 +59,20 @@ npm install palindrom --save
...
 ```
 
-###### Or just download it manually from [github](https://github.com/Palindrom/Palindrom/archive/master.zip).
+Note: The NPM package uses an entry point without a default export. Depending on your environment (Node or browser), you should choose one of the two named exports available: `Palindrom` and `PalindromDOM`.
```
tomalec: I'd say "you could" as I see reasonable use cases to use both in browser.

README.md
warpech: In my mind these are two different examples, all you ever need is either Palindrom or PalindromDOM. Should I present it better?

tomalec: no, it's fine as it is, that was just a question.

README.md
warpech: I deleted this whole sentence :)

warpech:
I made few changes that I think make it very clear which named export to use. I think that the lack of default export does not require explanation.

PR #269, opened 2019-09-02
bugfix: remove throwing an error if new remoteURL is different from the prevoious one

### LINKED ISSUE #268: Regarding setting the X-Location header on application/json-patch+json requests from PalindromClient
I'm not sure exactly what the problem is here, but if the server-side Palindrom implementation sets the `X-Location` header in a response to a HTTP request for `application/json-patch+json` (which, for example, happens when requesting a new URL through a link), we get the following error in the browser:

```
Error: Session lost.
Server replied with a different session ID than the already set one.
Possibly a server restart happened while you were working.
Please reload the page.
Previous session ID: https://host-only:5001/palindrom/acd0a6c2371b4246b446d9a35fe154fe
New session ID: /palindrom/acd0a6c2371b4246b446d9a35fe154fe
    at k.setRemoteUrl (palindrom-dom.min.js:14)
    at k.handleResponseHeader (palindrom-dom.min.js:14)
    at palindrom-dom.min.js:14
```

If we do not set the header at all, things work as they're supposed to.

cc. @miyconst 

With `X-Location` header in response:

Screen Shot 2019-08-21 at 17 08 38

No `X-Location` header in response:

Screen Shot 2019-08-21 at 17 10 08

warpech:
Fixes #268

See the commit message for description.


miyconst:
@warpech please also update readme.

warpech:
> @warpech please also update readme.

@miyconst I think this topic was not covered in the docs at all. However, I have just pushed an explanation of the current behavior,

Worth noting: there is no test, because:

1. This check was actually not covered in tests
2. We have another WIP branch with many changes and simplifications in the testing framework: https://github.com/Palindrom/Palindrom/tree/wip-modules

@miyconst or @erikvk I would be grateful if you could test it manually with the app. Then I will make a release of Palindrom.

warpech:
Closing because I am no longer involved in the project. If anyone wants to continue, you may open a new PR from my branch

PR #275, opened 2019-09-27
YAGNI, Remove support for single versioned JSON Patch communications,

tomalec:
aka `JSONPatchQueueSynchronous` - scenarios, when whole JSON object needs to be frozen after every local change, until the remote responds.

@miyconst, @eriksunsol I guess you are not using Palindrom for such scenarios, are you?


miyconst:
@tomalec I don't remember a single case where this feature was actually used, but let's see what @eriksunsol says.

tomalec:
I'm merging it, we will still have time to revert it before the next release.

eriksunsol:
Sorry, I didn't see this until now. No we don't use this scenario that I'm aware.

